PR #11, opened 2025-08-26
Update Makefile

tantive:
make run should also install the required dependencies


andig:
This should be make install

PR #13, opened 2025-08-31
restructured python code

ekkea:
- moved optimizer code to a separate file
- improved documentation and structure
- no functional changes
- removed example data from py file


evccoptimizer.py
```diff
@@ -0,0 +1,299 @@
+import pulp
```
andig, line 1: Can we rename this to optimizer? It‘s really agnostic of evcc. If example is removed, we‘d also need to update openapi and regenerate the go client.

ekkea: sure. I thought the name should not be totally generic as it has a specific purpose. We could also call it EmOptimizer (energy management) ... whatever you prefer.
And yes, you are right, I'll take care of the interface, I forgot to look at that.

ekkea: done.

ekkea: will do it in another PR

PR #19, opened 2025-09-16
Fix strategy incentive and unintended discharging to grid

ekkea:
fixed:
- promoted s[i][t] instead of c[i][t] which looks right in many cases but creates weird results in others.
- effectless lock against discharging to grid
- charging against discharging not explicitely locked

### DIFF
--- a/optimizer.py
+++ b/optimizer.py
@@ -124,7 +124,7 @@ class Optimizer:
             else:
                 self.variables['z_c'][i] = None
 
-        # Binary variable to lock chrging against discharging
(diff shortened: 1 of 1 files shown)


optimizer.py
```diff
@@ -124,6 +124,14 @@ def _setup_variables(self):
             else:
                 self.variables['z_c'][i] = None
 
+        # Binary variable to lock chrging against discharging
```
andig, line 127:
```suggestion
        # Binary variable to lock charging against discharging
```

PR #25, opened 2025-09-23
added API proposal for various electrical loads

ekkea:
- profiled: The load is determined by a load profile (average power per time step) for a sequience 
    of time steps. The profile may be shorter of equal to the length of the model's time horizon.
    In this case, the optimizer will determin the best time to start at or after t_start. However,
    the placement will be such that the profile will be completed at or before the end of the 
    model's time horizon.
    Typical use cases: constant loads like bitcoin miners, loads with a defined profile like wahsing machines.
- buffered: The load is flexible over time, typically feeding a thermal buffer like a hot water tank or a 
    heating system. The load is defined by an energy amount in Wh to be delivered within the model's time horizon.
    Along with that, a buffer size in Wh can be specified to to provide a sufficient distribution of the load drawn
    over time. This constraint will make sure that the integral of the optimized load profile will never deviate 
    from the average load by more than 1/2 of the buffer size in either direction.
- value_driven: The load can be fully controlled by the optimization within p_min and p_max, using its
    indicated value per Wh.


ekkea:
we can discuss this. I would assume that load profiles would rather be computed in evcc, even if they require a model to compute them (in this case it would be a simple thermal building model based on a transfer function or so).
On the other hand, an argument for implementing such things on the python side is that there is an excellent library support for all kind of data processing :thinking:  
@andig , what do you think?

lfloeer:
I also like your idea of making evcc the source of truth. I could imagine that one could add a plugin API to provide load profiles, similar to how tariffs work. If evcc would also record the historic data, there would be nice options for training individual forecasting models on the historic data.

ekkea:
I agree, that is an argument to have it there.

PR #31, opened 2025-10-04
add static grid limit as load management restriction

### LINKED ISSUE #23: Add static grid limit as load management restriction
If grid connection has a load management limit (e.g. 4.2kW), optimizer should not draw more than that from grid.

ekkea:
Fix #23

- added optional limits for grid import and export to interface
- added optional demand rate parameter for excess power beyond the import limit
- both limits accept violation if there is no physical solution otherwise. See also discussion below.

TO BE DISCUSSED:
- how shall we deal with cases where the import limit gets violated?
  - return infeasible 
  - return a specific flag
  - return a time series of the excess power drawn
  - reduce demanded charging first, then one of the above options
- how shall we handle the case that a given export limit gets violated (typically caused by high solar yield)
  - consider that in those cases the solar power is reduced to match the limit. Always return a time series giving the actual solar power produced (ft - p_exp_pen).
  - return infeasible
  - ???

### DIFF
--- a/cmd/client.go
+++ b/cmd/client.go
@@ -118,40 +118,9 @@ func main() {
 	}
 
 	res := *resp.JSON200
-	b, _ := json.MarshalIndent(res, "", "  ")
-
-	// Write request and response into a file that can serve as a test case description
-	// Create the ./tmp directory if it does not exist
-	{
-		test_items := map[string]any{}
-		combined := map[string]interface{}{
-			"request":           req,
-			"expected_response": res,
-			"test_items":        test_items,
-		}
-		cb, err := json.MarshalIndent(combined, "", "  ")
-		if err != nil {
-			log.Fatal(err)
-		}
-
-		// Ensure tmp directory exists
-		if _, err := os.Stat("./tmp"); err != nil {
-			if os.IsNotExist(err) {
-				if err := os.Mkdir("tmp", 0755); err != nil {
-					log.Fatal(err)
-				}
-			} else {
-				log.Fatal(err)
-			}
-		}
-
-		// Write combined JSON to file
-		if err := os.WriteFile("tmp/test_case.json", cb, 0666); err != nil {
-			log.Fatal(err)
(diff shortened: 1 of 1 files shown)


cmd/client.go
```diff
@@ -118,8 +118,40 @@ func main() {
 	}
 
 	res := *resp.JSON200
+	b, _ := json.MarshalIndent(res, "", "  ")
+
+	// Write request and response into a file that can serve as a test case description
+	// Create the ./tmp directory if it does not exist
+	{
+		test_items := map[string]any{}
+		combined := map[string]interface{}{
```
andig:
```suggestion
		combined := map[string]any{
```

cmd/client.go
```diff
@@ -118,8 +118,40 @@ func main() {
 	}
 
 	res := *resp.JSON200
+	b, _ := json.MarshalIndent(res, "", "  ")
```
andig: seems this should be behind a command line option

cmd/client.go
ekkea: You'll know better, I took AI to help me with the Go stuff.

cmd/client.go
```diff
@@ -118,8 +118,40 @@ func main() {
...
+			"expected_response": res,
+			"test_items":        test_items,
+		}
+		cb, err := json.MarshalIndent(combined, "", "  ")
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		// Ensure tmp directory exists
+		if _, err := os.Stat("./tmp"); err != nil {
```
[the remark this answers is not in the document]
ekkea: Fine for me, I'm a Go rookie, so I don't know ;-)

ekkea:
I'll take care of the changes later tonight

cmd/client.go
andig: Ich hab den Testexport erstmal rausgeworfen- ist ein separates Thema.

PR #32, opened 2025-10-08
fix formatting

lfloeer:


andig:
Another inconsistency? Applying autopep8 locally, this is not required. It only happens when running make. Question is why and where is the inconsistency?

lfloeer:
Now extension, make, and CI should be consistent.

PR #44, opened 2025-11-22
14 add test case editor

ekkea:
improved test script and added tool to generate , update and run+visualize test cases.


ekkea:
Good to merge. There are already a few test cases in ./test_cases, they are still valid. No new test cases here. This PR is on a tool that allows to rerun test cases after a code change and analyse the differences  to the previous results (in case there are any). The tool also allows to update test cases (set the current results as expected results) or create new ones.
I can add a readme if you think that makes sense, otherwise this is complete.

PR #47, opened 2026-01-31
Add validation errors to OpenAPI spec

t0mas:
Update the OpenAPI spec and go client to include validation errors from the @api.expect annotations. The error responses look like this:
```
{
    "errors": {
        "batteries.0.c_max": "'c_max' is a required property",
        "batteries.0.c_min": "'c_min' is a required property",
        "batteries.0.d_max": "'d_max' is a required property",
        "batteries.0.p_a": "'p_a' is a required property",
        "batteries.0.s_initial": "'s_initial' is a required property",
        "batteries.0.s_max": "'s_max' is a required property",
        "batteries.0.s_min": "'s_min' is a required property",
        "time_series.dt": "'dt' is a required property",
        "time_series.gt": "'gt' is a required property",
        "time_series.p_E": "'p_E' is a required property",
        "time_series.p_N": "'p_N' is a required property"
    },
    "message": "Input payload validation failed"
}
```

Will use this on the evcc side to better log the errors. See https://github.com/evcc-io/evcc/issues/27133

### DIFF
--- a/client/client.gen.go
+++ b/client/client.gen.go
@@ -101,8 +101,8 @@ type BatteryResult struct {
 
 // Error defines model for Error.
 type Error struct {
-	// Errors Field-specific validation errors. Keys are field paths (e.g., "batteries.0.s_max"), values are error messages.
-	Errors map[string]string `json:"errors,omitempty"`
+	// Details Field-specific validation errors. Keys are field paths (e.g., "batteries.0.s_max"), values are error messages.
+	Details map[string]string `json:"details,omitempty"`
 
 	// Message Error message describing what went wrong
 	Message string `json:"message,omitempty"`


t0mas:
The framework automatically reports them, adding them to the spec makes it easier on the evcc side to report them which is what I was working on.

openapi.yaml
```diff
@@ -453,7 +453,15 @@ components:
         message:
           type: string
           description: Error message describing what went wrong
-          example: "All time series must have the same length"
+          example: "Input payload validation failed"
+        errors:
```
ekkea: an `errors `field in an `Error `object may be a bit confusing. Can we name it `details` or so?

ekkea:
I'd say that more detailed reporting that helps trouble shooting is a helpful improvement. By nature you'll only need it when you run into a problem of course. However, this way we could also return more helpful statements if the optimization fails for non obvious reasons (e.g. over-constraints) using the same mechanism. I'd support the change.

openapi.yaml
andig: Why not just an error string as top level property? The combination of map and string is quite confusing.

t0mas: We could rename it or change it to a string by overriding the default error/validation behavior of the framework. Adding a handler via @api.errorhandler(BadRequest) in app.py should be enough to then define our own message and format.

I think the structured "details" field based on what the framework already generates is the most flexible approach. We can then later add more validation annotations and messages without writing a lot of specific error message code.

PR #53, opened 2026-03-02
Don't report micro violations of grid limits

RenatusRo:
Just a suggestion and any other value than 0.1 might be better.
but currently resonses appear like 
"grid_import_overshoot": [0, 0, 0, 1.3600118e-10, 0, 0, 1.3600118e-10,...


ekkea:
of course we can clip that on the optimizer end, however, I would rather leave it up to the user of the data to do that. Floating point numbers are as good as never exact and they may change slightly through casting / encoding / conversion anyway.

PR #58, opened 2026-03-13
Add absolute physical grid import limit p_max_abs_imp

RenatusRo:


SourceryAI (commented):
Hey - I've found 2 issues, and left some high level feedback:

- Consider updating the `GridConfig` dataclass type for `p_max_abs_imp` to be optional (e.g. `Optional[float]`) to match how it is used (`None` default and `is not None` checks), and to avoid confusing type checkers.
- It might be worth adding input validation to ensure `p_max_abs_imp >= p_max_imp`, since otherwise the constraint `p_max_imp_exc <= p_max_abs_imp - p_max_imp` can become negative and make the model infeasible.

<details>
<summary>Prompt for AI Agents</summary>

~~~markdown
Please address the comments from this code review:

## Overall Comments
- Consider updating the `GridConfig` dataclass type for `p_max_abs_imp` to be optional (e.g. `Optional[float]`) to match how it is used (`None` default and `is not None` checks), and to avoid confusing type checkers.
- It might be worth adding input validation to ensure `p_max_abs_imp >= p_max_imp`, since otherwise the constraint `p_max_imp_exc <= p_max_abs_imp - p_max_imp` can become negative and make the model infeasible.

## Individual Comments

### Comment 1
<location path="src/optimizer/optimizer.py" line_range="18-20" />
<code_context>
 @dataclass
 class GridConfig:
     p_max_imp: float
+    p_max_abs_imp: float
     p_max_exp: float
     prc_p_exc_imp: float
</code_context>
<issue_to_address>
**suggestion:** Consider making `p_max_abs_imp` (and possibly `p_max_imp`) explicitly optional since they can be `None` in practice.

In `app.py`, `GridConfig` is constructed with `grid_data.get(..., None)` for these fields, so they may be `None` at runtime despite being annotated as `float`. This can mislead type checkers and hide `None`-handling bugs. Please either annotate them as `Optional[float]` (and do the same for other nullable fields) or ensure `None` is never passed in.

Suggested implementation:

```python
from dataclasses import dataclass
from typing import Optional

```

```python
@dataclass
class GridConfig:
    p_max_imp: Optional[float]
    p_max_abs_imp: Optional[float]
    p_max_exp: float
    prc_p_exc_imp: float

```

```python
            # hard physical cap on how far above p_max_imp the import can go
            if self.grid.p_max_abs_imp is not None and self.grid.p_max_imp is not None:
                self.problem += self.variables['p_max_imp_exc'] \
                    <= self.grid.p_max_abs_imp - self.grid.p_max_imp

```

Other configuration fields that can be `None` in practice and are currently typed as `float` (or other non-Optional types) should also be updated to `Optional[...]` in `GridConfig` and any similar config dataclasses. Any usage sites that perform arithmetic or comparisons on these fields should either:
1. Be guarded with `is not None` checks (as done above), or
2. Enforce non-nullability earlier (e.g. validate and coerce defaults when constructing `GridConfig` instead of passing `None` through).
You may also want to adjust the construction in `app.py` to avoid using `get(..., None)` where a value is required, or add explicit validation if `None` is not acceptable.
</issue_to_address>

### Comment 2
<location path="src/optimizer/optimizer.py" line_range="396-398" />
<code_context>
                 self.problem += self.variables['e_imp_lim_exc'][t] \
                     <= self.variables['p_max_imp_exc'] * self.time_series.dt[t] / 3600

+            # hard physical cap on how far above p_max_imp the import can go
+            if self.grid.p_max_abs_imp is not None:
+                self.problem += self.variables['p_max_imp_exc'] \
+                    <= self.grid.p_max_abs_imp - self.grid.p_max_imp
+
</code_context>
<issue_to_address>
**issue (bug_risk):** Guard against inconsistent configurations where `p_max_abs_imp` is set but `p_max_imp` is `None` or larger than `p_max_abs_imp`.

This constraint assumes both values are set and that `p_max_abs_imp >= p_max_imp`. If `p_max_imp` is `None`, model construction will fail; if `p_max_abs_imp < p_max_imp`, the RHS is negative and can silently cause infeasibility if `p_max_imp_exc` is constrained non-negative. Please either validate the configuration up front (e.g., require `p_max_imp` when `p_max_abs_imp` is set and enforce `p_max_abs_imp >= p_max_imp`), or adjust the formulation to handle the “only `p_max_abs_imp` set” case by constraining the absolute import variable instead of the excess variable.
</issue_to_address>
~~~

</details>

***

Hi @RenatusRo! 👋

Thanks for trying out Sourcery by commenting with `@sourcery-ai review`! 🚀

Install the [sourcery-ai bot](https://github.com/apps/sourcery-ai/installations/new) to get automatic code reviews on every pull request ✨

<sub>
Help me be more useful! Please click 👍 or 👎 on each comment and I'll use the feedback to improve your reviews.
</sub>

src/optimizer/optimizer.py
```diff
@@ -17,6 +17,7 @@ class OptimizationStrategy:
 @dataclass
 class GridConfig:
     p_max_imp: float
+    p_max_abs_imp: float
```
SourceryAI: **suggestion:** Consider making `p_max_abs_imp` (and possibly `p_max_imp`) explicitly optional since they can be `None` in practice.

In `app.py`, `GridConfig` is constructed with `grid_data.get(..., None)` for these fields, so they may be `None` at runtime despite being annotated as `float`. This can mislead type checkers and hide `None`-handling bugs. Please either annotate them as `Optional[float]` (and do the same for other nullable fields) or ensure `None` is never passed in.

Suggested implementation:

```python
from dataclasses import dataclass
from typing import Optional

```

```python
@dataclass
class GridConfig:
    p_max_imp: Optional[float]
    p_max_abs_imp: Optional[float]
    p_max_exp: float
    prc_p_exc_imp: float

```

```python
            # hard physical cap on how far above p_max_imp the import can go
            if self.grid.p_max_abs_imp is not None and self.grid.p_max_imp is not None:
                self.problem += self.variables['p_max_imp_exc'] \
                    <= self.grid.p_max_abs_imp - self.grid.p_max_imp

```

Other configuration fields that can be `None` in practice and are currently typed as `float` (or other non-Optional types) should also be updated to `Optional[...]` in `GridConfig` and any similar config dataclasses. Any usage sites that perform arithmetic or comparisons on these fields should either:
1. Be guarded with `is not None` checks (as done above), or
2. Enforce non-nullability earlier (e.g. validate and coerce defaults when constructing `GridConfig` instead of passing `None` through).
You may also want to adjust the construction in `app.py` to avoid using `get(..., None)` where a value is required, or add explicit validation if `None` is not acceptable.

src/optimizer/optimizer.py
```diff
@@ -392,6 +393,11 @@ def _add_energy_balance_constraints(self):
                 self.problem += self.variables['e_imp_lim_exc'][t] \
                     <= self.variables['p_max_imp_exc'] * self.time_series.dt[t] / 3600
 
+            # hard physical cap on how far above p_max_imp the import can go
+            if self.grid.p_max_abs_imp is not None:
+                self.problem += self.variables['p_max_imp_exc'] \
```
SourceryAI: **issue (bug_risk):** Guard against inconsistent configurations where `p_max_abs_imp` is set but `p_max_imp` is `None` or larger than `p_max_abs_imp`.

This constraint assumes both values are set and that `p_max_abs_imp >= p_max_imp`. If `p_max_imp` is `None`, model construction will fail; if `p_max_abs_imp < p_max_imp`, the RHS is negative and can silently cause infeasibility if `p_max_imp_exc` is constrained non-negative. Please either validate the configuration up front (e.g., require `p_max_imp` when `p_max_abs_imp` is set and enforce `p_max_abs_imp >= p_max_imp`), or adjust the formulation to handle the “only `p_max_abs_imp` set” case by constraining the absolute import variable instead of the excess variable.

andig:
It's unclear what this PR does :(

ekkea:
@RenatusRo , this sort of violates the approach we took in how the optimizer shall handle over-constraints that would end up in "`infeasible`" results. If the caller gets an `infeasible` back from the optimizer, he can hardly figure out what caused the over constraint. He can assume something but proving the assumption is not easy.
For this reason, such cases get handled by making the constraints flexible and then reporting the level of violation of the given limits, so that analysis is easy for the caller.
I don't see why we should go back from this.

RenatusRo:
Contemplated (part time) for weeks on this.
Thanks for the explanation, makes sense and is a clever solution for handling infeasibilities.
So, I rest my case. 

But I guess the issue will return via user requests, as there is currently no model for handling capacity
tariffs where energy cost is linked to some maximum power consumption metric. 
Probably this is hard to include in the short time horizon optimization anyway.
So in practice it might be sufficient to stick to the currently used heuristic: aim for some acceptable upper capacity limit and feed this to the optimizer.

That being said: that PR worked pretty well for me. 
And to illustrate the relevance for Flemish users:
Charging from the grid once in a month for 15 minutes at 22 kW puts an additional 80 Euro on your energy bill for that month (compared to 2 kW).
So in this real world context there is a difference between a hard physical upper limit and an economical soft limit with (feasable) penalized violations.

PR #77, opened 2026-05-31
fix: use max() for penalty scaling floor (closes #75)

### LINKED ISSUE #75: penalty limit logic is incorrect (min() called instead of max())
The bug has not surfaced in any report but should be fixed. Along with that, clean up the calculation of incentives.

andig:
## Summary

Fixes #75. The penalty-scaling base was computed with `np.min([self.max_import_price, 0.1e-3])`, despite the comment stating the intent was to keep the penalty positive. `min()` does the opposite:

- **Negative market prices** (which occur in real energy markets) drive the base negative, turning penalties into *rewards* in the maximizing objective — the optimizer would then seek out goal/limit violations.
- **Normal prices** above the `1e-4` floor get clamped down to that tiny floor, making penalties too weak to reliably dominate real costs.

The intended behavior is a **positive floor that scales with the highest import price**, which is `np.max()`.

## Changes

- Switch `np.min` → `np.max` for the penalty base.
- Deduplicate the expression — repeated 5× across the goal/SOC/grid-import/grid-export penalties — into a single `penalty_base` computed once (the "clean up the calculation of incentives" part of the issue).

Behavior is unchanged for typical positive prices above the floor, and corrected for the zero/negative-price edge cases the issue flagged.

## Testing

- `uv run pytest` → 13 passed
- `uv run ruff check` → clean

### DIFF
--- a/src/optimizer/optimizer.py
+++ b/src/optimizer/optimizer.py
@@ -81,11 +81,7 @@ class Optimizer:
         self.min_import_price = np.min(self.time_series.p_N)
         self.max_import_price = np.max(self.time_series.p_N)
 
-        # Base scaling for all penalty parameters. Penalties must stay strictly
-        # positive so that, in the maximizing objective, they actually discourage
-        # the violations they are subtracted for. Scale with the highest import
-        # price but never below a small positive floor, otherwise zero or negative
-        # market prices would weaken or even invert the penalties.
+        # scaling base for penalty parameters. Make sure goal_penalty is always positive.
         penalty_base = np.max([self.max_import_price, 0.1e-3])
 
         # scaling for penalty parameters
--- a/src/optimizer/optimizer.py
+++ b/src/optimizer/optimizer.py
@@ -81,17 +81,24 @@ class Optimizer:
         self.min_import_price = np.min(self.time_series.p_N)
         self.max_import_price = np.max(self.time_series.p_N)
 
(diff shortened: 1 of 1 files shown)


src/optimizer/optimizer.py
```diff
@@ -81,17 +81,24 @@ def __init__(self, strategy: OptimizationStrategy, grid: GridConfig, batteries:
...
 
-        # scaling for penalty parameters. Make sure goal_penalty is always positive
-        self.prc_e_goal_pen = np.min([self.max_import_price, 0.1e-3]) * 10e1
-        self.prc_p_goal_pen = np.min([self.max_import_price, 0.1e-3]) * np.max(self.time_series.dt) / 3600 * 10e1
-        self.prc_soc_exc_pen = np.min([self.max_import_price, 0.1e-3]) * 10e2
+        # Base scaling for all penalty parameters. Penalties must stay strictly
+        # positive so that, in the maximizing objective, they actually discourage
+        # the violations they are subtracted for. Scale with the highest import
+        # price but never below a small positive floor, otherwise zero or negative
+        # market prices would weaken or even invert the penalties.
```
andig:
```suggestion
        # scaling base for penalty parameters. Make sure goal_penalty is always positive.
```

andig:
@ekkea lgtm?

ekkea (approved):
Looks good, let's merge!

PR #88, opened 2026-07-15
perf: tighten indicator big-Ms to natural per-step bounds

andig:
Redone on top of main after #95 rewrote the charge demand block and #102 added the ramp penalty. The old two disjunct tightening no longer applies, the new indicators are tightened instead.

Every indicator constraint used the global `M = 1e6` while the largest per step energy in any test case is a few kWh. The relaxation is then nearly vacuous around the binaries: a fractional `z_p_demand` of ~0.003 satisfies the charge demand disjunction while paying neither the charge cost nor the penalty, and the flow direction indicators carry no information at the root. Each M becomes the smallest constant that keeps the inactive branch vacuous, so the integer feasible set is unchanged and the relaxation is strictly tighter.

- `s_max` for the two state of charge gap indicators, `z_p_demand` and `z_s_max_reached`. The state of charge has a lower bound of 0, so `s_max` is the largest value the left side can take
- the per step `c_min` energy for the row that keeps charging out of the band between 0 and `c_min`
- the per step charge and discharge caps for `z_c`, `z_cd` and the grid coupling rows, these are the variables' own upper bounds
- the limit threshold itself for `z_imp_lim` and `z_exp_lim`
- solar plus discharge to grid capacity, and demand plus grid charge capacity, for the flow direction indicators. Only when a grid limit is active does the opposite side's unbounded excess variable enter the balance, and only then does the global M remain
- cases 013 and 019 land on a different point of the same optimal plateau. The full objective is equal to 5.5e-7, only the penalty free `objective_value` decomposition differs, so both expected values are updated

Full suite, median of 5 runs per case, in process through the Flask test client, milliseconds:

| case | main | PR | delta |
|---|---|---|---|
| 009-discharge-before-import | 84.7 | 89.5 | 1.06x |
| 010-infesible-charge-goal | 34.8 | 35.6 | 1.02x |
| 011-infeasible-charge-demand | 82.8 | 69.7 | 0.84x |
| 012-early-charging-not-perfect | 82.2 | 108.8 | 1.32x |
| 013-grid-export-limit-hit | 64.4 | 58.5 | 0.91x |
| 014-grid-import-limit-violation | 42.0 | 45.8 | 1.09x |
| 015-low-soc-initial | 42.3 | 43.4 | 1.03x |
| 016-battery-charge-priotization-1 | 196.7 | 97.9 | 0.50x |
| 017-battery-charge-priotization-2 | 94.7 | 77.1 | 0.81x |
| 018-high-soc-initial | 93.3 | 101.1 | 1.08x |
| 019-unexpected-charge-spikes | 42.8 | 41.5 | 0.97x |
| **020-weird-charging-at-night** | **2422.0** | **668.7** | **0.28x** |
| **021-min-pv-use-case-with-weird-behavior** | **735.5** | **299.4** | **0.41x** |
| 022-feed-in-cap-recovery | 24.8 | 20.2 | 0.81x |
| 023-c_min-limit-kept-with-p_demand-set | 164.6 | 158.0 | 0.96x |
| 024-attenuate-demand-peaks | 25.6 | 20.6 | 0.80x |
| 025-attenuate-feedin-peaks | 22.3 | 19.3 | 0.87x |
| 026-attenuate-grid-peaks | 28.9 | 24.9 | 0.86x |
| 027-attenuate-feedin-peaks-below-cap | 23.0 | 20.4 | 0.89x |
| **TOTAL** | **4307.4** | **2000.4** | **0.46x** |

021 was the one regression in the earlier version of this branch, 0.14s to 0.38s. With the ramp penalty from #102 in main it improves instead. `pytest` wall clock goes from 9.2s to 2.2s. All cases stay Optimal, and the four strict peak shaping cases pass with unchanged schedules.

### DIFF
--- a/src/optimizer/optimizer.py
+++ b/src/optimizer/optimizer.py
@@ -408,12 +408,22 @@ class Optimizer:
                              == e_grid_exp
                              + self.time_series.gt[t])
 
-        # Constraints (4)-(5): Grid flow direction
+        # Constraints (4)-(5): Grid flow direction. Export/import have natural
+        # per-step caps (export: solar plus discharge-to-grid capacity; import:
+        # demand plus grid-charge capacity) that tighten the LP relaxation vs
+        # the global big-M without excluding any integer point. When a grid
+        # limit is active, the opposite side's unbounded excess variable enters
+        # the energy balance, so only then fall back to the global big-M.
+        cap_d_exp = sum(b.d_max for b in self.batteries if b.discharge_to_grid)
+        cap_c_imp = sum(b.c_max for b in self.batteries if b.charge_from_grid)
         for t in self.time_steps:
+            dth = self.time_series.dt[t] / 3600.
+            m_exp = self.time_series.ft[t] + cap_d_exp * dth if self.grid.p_max_imp is None else self.M
+            m_imp = self.time_series.gt[t] + cap_c_imp * dth if self.grid.p_max_exp is None else self.M
             # Export constraint
-            self.problem += self.variables['e'][t] <= self.M * self.variables['y'][t]
+            self.problem += self.variables['e'][t] <= m_exp * self.variables['y'][t]
             # Import constraint
-            self.problem += self.variables['n'][t] <= self.M * (1 - self.variables['y'][t])
+            self.problem += self.variables['n'][t] <= m_imp * (1 - self.variables['y'][t])
 
         # limit regular grid import power
         if self.grid.p_max_imp is not None:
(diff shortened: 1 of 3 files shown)


andig:
@ekkea found this during model optimization when porting CBC to Go.

andig:
@copilot resolve the merge conflicts in this pull request

andig:
@ekkea would appreciate best practice review of this one- entirely over my head and I can only trust Claude and measurements here

ekkea:
I think scaling M more carefully to the specific cases makes sense, everything is well done. - it is interesting to see where it improves the performance.

However, adding more complexity to the optimizer code will require a decent restructuring of the code to maintain it readable for human beings (I don't think Claude cares so much) at some point in time. Anyway, we have to get through the open topics first ...

PR #89, opened 2026-07-15
perf: skip charge/discharge lock binaries for one-directional batteries

andig:
## Full-suite benchmark (bundled CBC, single-thread, app-faithful build, median of 3)

Measured against #88 (this PR's base):

| case | #88 nodes/iters/wall | this PR nodes/iters/wall |
|---|---|---|
| 013 | 0 / 12 / 0.05s | 0 / 5 / 0.05s |
| 017 | 0 / 98 / 0.06s | 0 / 87 / 0.06s |
| 020 | 18 / 1442 / 0.62s | 18 / 1806 / 0.56s |
| **021** | **10 / 220 / 0.38s** | **52 / 798 / 1.24s** |
| others | unchanged | unchanged |
| **TOTAL** | **1.70s** | **2.50s** |

⚠ Honest tradeoff: the binary removal is a pure model-size win (−154 binaries on 020) and slightly helps 013/017/020, but **021 regresses 5×** (10→52 nodes) — with the vacuous rows gone, CBC's dive heuristics land elsewhere on 021's alternate-optima plateau and the suite total goes net-negative on this machine. The removal is structurally sound (optimum unchanged); whether smaller models or this machine's search-path luck matters more is a maintainer call. Happy to close if the wall-clock regression outweighs the size win.

PR #90, opened 2026-07-16
perf: scale solver objective 1e4; hard SOC bounds for in-range starts

andig:
Stacked on #88 (base `model/tight-big-m`; merge that first). Independent of #89.

Two changes, validated together against the full golden suite:

**1. Solver-facing objective ×1e4** (one line). The optimum sits at ~0.5–18 currency units with tie-break coefficients down at 1e-8; CBC's absolute tolerances (integrality 1e-6, cutoff increment 1e-5) can't separate near-tie nodes, so the search wanders the alternate-optima plateau. Scaling lets the tolerances bite. `get_clean_objective_value` recomputes from variable values, so reported numbers stay unscaled.

**2. Hard SOC bounds when `s_min ≤ s_initial ≤ s_max`.** The soft-penalty construction (2 penalty columns + 2 rows per battery per time step) is provably dead weight for in-range starts: the penalty (0.27/Wh) exceeds any achievable gain (<3e-4/Wh) by three orders of magnitude, so penalties are zero at every optimum and `[s_min, s_max]` can live on the variable bounds. The soft path stays for out-of-range starts (015/018 exercise it).

## Full-suite benchmark (bundled CBC, single-thread, app-faithful build, median of 3)

| case | main | #88 (base) | this PR |
|---|---|---|---|
| 009 | 0nd / 160it / 0.08s | 0 / 121 / 0.09s | 0 / 129 / 0.08s |
| 016 | 2 / 1256 / 0.18s | 0 / 148 / 0.09s | 2 / 189 / 0.08s |
| 017 | 4 / 97 / 0.09s | 0 / 98 / 0.06s | 0 / 0 / 0.04s |
| 018 | 8 / 546 / 0.09s | 2 / 266 / 0.09s | 2 / 261 / 0.09s |
| 020 | 86 / 5233 / 1.02s | 18 / 1442 / 0.62s | 24 / 1714 / 0.61s |
| **021** | 0 / 15 / 0.14s | 10 / 220 / 0.38s | **0 / 0 / 0.09s** |
| others | ≤0.05s, 0 nodes | unchanged | unchanged |
| **TOTAL** | **1.93s** | **1.70s** | **1.36s** |

Every case is equal-or-better than **main** — notably this heals the base branch's 021 drift (10 nodes → root-solved, faster than main ever was), and side effects each change shows alone (SOC bounds alone push 009 to 770 iterations) cancel under scaling. Stable across reruns.

**Equivalence:** full pulp objectives (÷1e4) match the base branch to ≤3e-6 absolute, 016/017/020/021 exact. The 013/019 golden `objective_value`s land on different points of the same optimal plateau (019 returns to main's original value); optimum unchanged.

### DIFF
--- a/src/optimizer/optimizer.py
+++ b/src/optimizer/optimizer.py
@@ -141,7 +141,10 @@ class Optimizer:
         self.variables['s'] = {}
         for i, bat in enumerate(self.batteries):
             self.variables['s'][i] = [
-                pulp.LpVariable(f"s_{i}_{t}", lowBound=0, upBound=bat.s_capacity)
+                pulp.LpVariable(
+                    f"s_{i}_{t}",
+                    lowBound=bat.s_min if bat.s_min <= bat.s_initial <= bat.s_max else 0,
+                    upBound=bat.s_max if bat.s_min <= bat.s_initial <= bat.s_max else bat.s_capacity)
                 for t in self.time_steps
             ]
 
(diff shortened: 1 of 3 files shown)


andig:
@ekkea found this during model optimization when porting CBC to Go.

PR #98, opened 2026-07-22
feat: add /optimize/level-schedule endpoint for solar-first strategy

1va13n7in:
Add /optimize/level-schedule endpoint for solar-first strategy.

Finds the optimal flat export ceiling that charges a battery from solar
surplus while minimizing peak grid export. Response uses the same format
as /optimize/charge-schedule for identical downstream processing.

Depends on #97 (charge_profile).

Part of evcc-io/evcc#31963


1va13n7in:
Production results from a 34 kWp / 42.58 kWh LiFePO4 system.

Solar first active -- flat export ceiling during charging:
2026-07-29_08-52-25-optimize-SOC-33


The grey line (grid power) stays flat while the battery charges.
CC-CV taper visible as charge power decreases toward end of window.

**Actual power flows (Home Assistant):**
2026-07-29-charging

Real-world result from the same day. Battery charged from solar
surplus, export stayed nearly flat during charging.

The evcc-side PR (strategy routing, UI) is ready and waiting
on this to be merged and tagged.

andig:
Its still unclear to me (can't find in which related issue I've asked but didn't get an answer):
- why would this depend on the knee PR
- what's the difference to [grid peak leveling](https://github.com/evcc-io/optimizer#levelling-grid-peaks)

1va13n7in:
The dependency on #97 is because find_optimal_level() calls
max_charge_power() from charging_profiles.py which #97 introduces.
Without the taper, the level algorithm would assume constant c_max
up to 100% SoC, producing unrealistic schedules. That said, it could
work without taper (just less accurate) if you prefer to decouple them.

The difference to attenuate_grid_peaks: that is a cost-neutral
tiebreaker inside the MILP. It nudges toward charging during high
solar but the primary objective is still cost minimization.

Solar-first is a different optimization objective entirely. It
directly computes the flat export ceiling that fills the battery
to target SoC. The export curve is mathematically flat by
construction, not an approximation through incentive weighting.

For flat-rate tariffs both point in the same direction, but
solar-first is purpose-built for the self-consumption goal.

CC-CVtaper_Solar-first

andig:
> For flat-rate tariffs both point in the same direction, but
solar-first is purpose-built for the self-consumption goal.

I still don't get why we should be doing this. Why would self-consumption be a relevant goal and preferable over economic value plus peak shaving?

/cc @premultiply

PR #101, opened 2026-07-24
test: cover peak shaping strategies with data driven cases

### LINKED ISSUE #100: Add peak shaving test cases
The implementation looks good and makes sense. A bit hard to read at the first glance but elegant.

However, I would suggest not to add specific test scripts for each feature add. That will make it hard to figure out what went wrong if one of the tests fails on a future feature add. I'd prefer to add test cases to the current test system. Or do I miss something here?

I can add appropriate test cases, but I cannot commit into this PR.

_Originally posted by @ekkea in https://github.com/evcc-io/optimizer/issues/96#issuecomment-5062912570_

andig:
Fix https://github.com/evcc-io/optimizer/pull/100, depends on https://github.com/evcc-io/optimizer/pull/102

Follows up on https://github.com/evcc-io/optimizer/pull/96#issuecomment-5062912570: the peak shaping coverage lives in the existing data driven harness instead of a feature specific test module.

- three test cases, one per strategy: `attenuate_demand_peaks` levels grid import, `attenuate_feedin_peaks` moves charging into the solar hump, `attenuate_grid_peaks` does both
- a test case may now set `"strict": true`, which compares grid import, grid export, and the battery schedules with `numpy.allclose` on top of status and objective value
- strict is needed here because peak shaping only decides between cost neutral schedules. Without it the three cases pass unchanged when peak shaping is removed, the objective value differs only in the eighth digit
- existing cases stay non strict, several of them have equally optimal alternatives whose schedule comparison would be arbitrary
- `testbench.py update` keeps the strict flag when it refreshes a case


andig:
@ekkea hope this is a better direction

andig:
Added `test_cases/026-attenuate-feedin-peaks-below-cap.json`.

The three strategy cases pin the absolute peak; this one pins the shape of the profile *below* the peak. The surplus of the first step is far beyond `c_max`, so the export peak is fixed there and the remaining 170 Wh can be charged anywhere in the following steps without changing it — a schedule that only caps the peak has a free choice there, and the strict comparison pins the leveled one.

It fails on current main and needs #102:

- without the ramp penalty: `grid_export [950, 50, 50, 50, 80]`
- with it: `grid_export [950, 57.5, 57.5, 57.5, 57.5]`

ekkea:
I agree, this approach makes sense, the assert on the target function value was only thought as the start anyway.
Good to merge to main from my perspective.

andig:
Performance test result, run on the branch after the renumbering (macOS, Python 3.13, CBC, gunicorn on 127.0.0.1:7050).

Per case latency, 5 sequential requests each, milliseconds:

```
case                                          median      min      max
009-discharge-before-import                       92       88      104
010-infesible-charge-goal                         41       39       43
011-infeasible-charge-demand                      86       85      103
012-early-charging-not-perfect                    89       81      102
013-grid-export-limit-hit                         65       57       74
014-grid-import-limit-violation                   58       41      202
015-low-soc-initial                               45       44       47
016-battery-charge-priotization-1                207      200      213
017-battery-charge-priotization-2                 92       87      116
018-high-soc-initial                              96       88       98
019-unexpected-charge-spikes                      47       46       48
020-weird-charging-at-night                     2451     2418     2470
021-min-pv-use-case-with-weird-behavior          720      714      732
022-feed-in-cap-recovery                          22       21       25
023-c_min-limit-kept-with-p_demand-set           162      156      166
024-attenuate-demand-peaks                        24       22       25
025-attenuate-feedin-peaks                        22       21       29
026-attenuate-grid-peaks                          29       25       32
027-attenuate-feedin-peaks-below-cap              22       21       22
```

The four new cases are the cheapest in the suite, 22 to 29 ms. They use five time steps, so they add roughly 100 ms to a full suite run that is dominated by `020-weird-charging-at-night` at 2.4 s. `make test` is 20 passed in 9.2 s.

`make loadtest`, 30 s per stage, random case per request:

```
                   reqs  fails    avg    min    max    med  p95    req/s
main,   2 users      79      0    717     42   4904    170  2700    2.74
branch, 2 users     110      0    444     40   2549    140  2400    4.40
main,   4 users      86      0   1247    162   5053    850  3400    2.91
branch, 4 users     117      0    831    110   5569    340  3800    4.05
```

The branch looks faster only because locust picks a random case per request and the four new fast cases dilute the mix from 15 to 19 scenarios. Server code is untouched by this PR, so the mix is the whole difference. The per case table above is the meaningful comparison, and no existing case moved.

Issue #114, opened 2026-07-25
Flat tariff requests exhaust the solver time limit: the cost neutral tie breakers are the bottleneck


andig:
Since #112 the access log records per request latency, and since #113 the requests that exhaust the solver are kept. Both are deployed. This is what they show.

About 2.6 percent of `POST /optimize/charge-schedule` calls run into `OPTIMIZER_TIME_LIMIT` and return whatever CBC has at 20 seconds. The p50 is around 0.5 s, so the p99 is not a slow median, it is a cliff: 495 of 19006 calls over a two hour window, and the p99 sits at 20.5 s in every five minute bucket.

## What the slow requests look like

293 captured requests, compared against the 19 golden cases as the only fast reference available:

| feature | slow median | golden median |
|---|---|---|
| distinct import prices, share of steps | 0.007 | 0.200 |
| longest run of identical prices | 141 | 5 |
| steps | 192 | 35 |
| headroom to `s_max`, share | 0.20 | 0.76 |
| first `dt` | 504 s | 1944 s |
| uniform `dt` | 0 % | 26 % |

192 of the 293 have exactly **one** distinct import price across the whole horizon, 216 have two or fewer, and the export price is single valued in 256. The typical shape is a fixed tariff, a 33 to 35 hour horizon at 900 s with a partial first interval, and two batteries that are already close to full.

Battery symmetry is not involved: only 10 percent have two or more identical batteries, and that subset is not slower.

## Why they are slow

The tie breakers, not the physics and not the horizon. Same captured request, same 7794 rows and 1732 integer variables, only `charging_strategy` swapped:

| strategy | solve |
|---|---|
| `none` | 0.39 s |
| `attenuate_demand_peaks` | 1.55 s |
| `attenuate_feedin_peaks` | 1.96 s |
| `attenuate_grid_peaks` | 7.90 s |
| `charge_before_export` | 28.60 s |

A 433 step request, 4.5 days of horizon, solves in 0.39 s with no strategy and takes 73 times as long with a cost neutral one. On a second request, 192 steps and 4 batteries, `none` is 4.90 s and `attenuate_grid_peaks` is 8.62 s.

[...]

PR #115, opened 2026-07-25
perf: decide strategy ties in a second solve, not with a weight

andig:
Fixes the main part of #114. Rebased onto #104, which changed what these settings are for.

#114 measured that the cost neutral tie breakers, not the physics and not the horizon, drive the requests that exhaust `OPTIMIZER_TIME_LIMIT`. Their coefficients sit inside the solver's tolerance band, so they cannot prune anything. On a flat tariff, where the cost optimum is a plateau of equally priced schedules, the search walks that plateau instead of deciding the tie. 4196 of 6245 captured slow requests carry exactly one distinct import price.

**This PR changed shape after review.** It was opened with an absolute MIP gap and a weight on the strategy terms. @ekkea objected that a gap which stops the search a cent short must also swallow the strategy terms, since they are orders below a cent, and asked for proof that this does not end in a loss of strategy. Measured, the objection holds and no weight fixes it, so the two are now solved in sequence instead. The old approach and the numbers behind it are in the review thread.

### Why the weight could not carry it

The total weighted preference objective, in currency, against `gap_abs = 0.01`, over the golden cases:

| case | \|cost\| | \|3 · preference\| | gap / preference |
|---|---|---|---|
| 018-high-soc-initial | 162318 | 0.000067 | **149×** |
| 009-discharge-before-import | 3.48 | 0.000214 | **47×** |
| 010-infesible-charge-goal | 1397 | 0.000949 | **10×** |
| 023-c_min-limit-kept-with-p_demand-set | 0.74 | 0.001360 | **7×** |
| 011-infeasible-charge-demand | 2.52 | 0.001554 | **6×** |
| 012-early-charging-not-perfect | 5.17 | 0.002758 | **4×** |
| 015-low-soc-initial | 5253 | 0.003435 | **3×** |
| 013, 024-027 (peak and ramp) | — | 0.073 - 4.73 | ≪ 1 |

On eight of nineteen the entire weighted preference objective is worth less than the gap. The preference terms are one signed, so their total magnitude bounds their spread: on those cases every strategy variant sits inside the acceptance window and the solver returns whichever incumbent it happens to hold. The weight buys a factor 3 against a shortfall of 3 to 150, and it cannot be raised, 10 already gives up 0.4 percent of real money on 023.

That is not a tuning problem. It is what an absolute gap on a single objective means.

### What it does now

`solve()` runs two solves over the same model.

1. maximize `cost_objective` with `OPTIMIZER_GAP_ABS`, one cent. Real money only, and this is where the speed comes from.
2. maximize `preference_objective` under a constraint that keeps the money the first stage found. This is where the strategies are decided.

Strategy adherence becomes something the model solves for rather than something that emerges from the ratio between coefficients, so `OPTIMIZER_STRATEGY_WEIGHT` and its cliff table are gone. `OPTIMIZER_PREFERENCE_BUDGET` takes its place, defaulting to zero, the cost neutrality the strategies are documented to have. It is not a knob to reach for: the second stage is indifferent to money below the bound, so every unit granted is spent, measured exactly at budgets of 1 and of 100.

Both stages share one wall clock. Deciding the tie to proven optimality turns out to be its own hard problem, as expensive as the cost optimum on exactly the requests this is meant to help, so the tie break gets a quarter of the limit rather than whatever is left of it, and what it does not finish is kept anyway: a tie break stopped by the clock still holds an incumbent, and the alternative is the first stage schedule, which is no tie break at all. The improvement and the money are verified in `_solve_preferences` rather than read off the solver status.

### Measured

335 unique requests from the slow request dump, every one of them a request that exhausted the limit in production, run at the production limit of 10 s:

| variant | median | p90 | p99 | over 10 s | strategy better than main | strategy worse |
|---|---|---|---|---|---|---|
| main | 7.65 s | 10.19 s | 10.59 s | 149 | — | — |
| weight 3 + gap, as this PR was opened | 0.72 s | 10.11 s | 10.29 s | 47 | 51 | **112** |
| **two stage** | **2.99 s** | 10.10 s | 10.38 s | **52** | **120** | 79 |

The variant this PR was opened with is behind main on strategy adherence on 112 of 335 requests, which is the reviewer's objection showing up on production data rather than on a constructed case. The two stage version keeps almost all of the speed and turns that column around.

Money on the requests where both runs converged, in currency: 15 of 175 worse, worst 0.0071, median -2e-6. The rows where either run hit the limit are not evidence in either direction, neither result is converged there.

Nothing in the golden set moves, the stored expectations are untouched.

### Three things the dump brought out, which the golden cases hid

**PuLP's `warmStart=True` makes CBC lie.** Handed the first stage solution as a MIP start, CBC returns a strictly worse schedule and reports it as proven optimal, and it declares the model infeasible over a cost bound that same start satisfies. On request `4f28656db0ba`, preference -0.806 warm against -0.610 cold, where the cold value matches a single joint solve to the last digit. Every "CBC presolve is broken" symptom chased here turned out to be this, so presolve stays at its default.

**CBC misses the cost bound by its feasibility tolerance**, like any other row, 1.2e-5 on a bound around 1263. A verification of that bound needs a tolerance of its own or it throws away good solutions, `COST_BOUND_TOLERANCE`.

**The peak strategies are not cost neutral.** Levelling burns round trip efficiency: `024-attenuate-demand-peaks` needs about 3.5e-5 currency to level, and refuses to level under a stricter bound. The claim in AGENTS.md that the strategies only decide between economically equal solutions holds for the tie breakers, not for the tier 3 peak and ramp terms. Worth a separate look.

### What this gives away

Per request, at the default budget of zero: `COST_BOUND_SLACK` plus `COST_BOUND_TOLERANCE`, so at most 1.1e-4 currency, a hundredth of a cent, and it is spent rather than merely allowed. Three orders below the cent the cost stage may already leave on the table. A third stage that maximizes cost again under a preference bound would reclaim it, at the price of a third solve, which did not look worth it.

The third review comment, on `OBJECTIVE_SCALE` being a constant, is not addressed here. It came in with #104 and stands on its own, so it gets its own PR.


andig:
@ekkea this should fix the p99 latency

andig:
I've had Claude analyze all current long-running (i.e. p99 10s) requests:

| type | pop % | slow on main | slow on PR 115 | of the slow ones, PR 115 fixes |
|---|---|---|---|---|
| `charge_before_export` / flat | 38.2% | 58% | 33% | 3 of 7 |
| `attenuate_grid_peaks` / flat | 22.3% | 33% | 0% | 4 of 4 |
| `attenuate_grid_peaks` / varying | 19.6% | 25% | 8% | 2 of 3 |
| `charge_before_export` / varying | 16.9% | 67% | 33% | 5 of 8 |
| `none` / flat | 2.6% | 58% | 17% | 5 of 7 |
| `none` / varying | 0.5% | 25% | 0% | 1 of 1 |

Population-weighted: 47.5% slow on main → 20.4% with #115

src/optimizer/optimizer.py
```diff
@@ -97,6 +97,9 @@ def __init__(self, strategy: OptimizationStrategy, grid: GridConfig, batteries:
         self.time_steps = range(self.T)
         # the optimization problem
         self.problem = None
+        # objective split, filled by _setup_target_function: real money and preferences
+        self.cost_objective = 0
```
ekkea, line 180: splitting the objectives into "real costs" and "preferences" may be a nice approach to improve the code readability but does not add anything new to the model and its structure. There may be arbitrary performance wins by slightly changing ratios between different objective contributions but this is not going to be deterministic over all cases.

src/optimizer/optimizer.py
```diff
@@ -613,13 +631,17 @@ def solve(self) -> Dict:
...
-            threads=self.settings.num_threads,
-            timeLimit=self.settings.time_limit,
-        )
         with TemporaryDirectory() as tmpdir:
+            # the absolute gap is in currency units, so it says "do not spend time on a difference
+            # worth less than this" instead of asking for proven optimality on a plateau of equal
+            # schedules. It bounds money only: the strategy terms are far smaller than any useful
+            # gap, so the tie breaking is left to their weight, not to where the search stops.
+            # the model handed to the solver is scaled by OBJECTIVE_SCALE, so the gap is too,
+            # otherwise a cent would read as a millionth of one.
```
ekkea: I cannot follow the logic here. If the optimizer will stop as soon as the cost optimum is reached by a gap of e.g. 1ct, that says, that it will ignore strategy objectives (being a few magnitudes smaller) as soon as they tie into the branch and bound algorithm (i.e. as soon as they depend on integer variables).
We would need to prove that this does not end up in an unwanted "loss of strategy".

src/optimizer/optimizer.py
ekkea: I'm not sure if it is smart to set OBJECTIVE_SCALE to a constant value. It may be a great choice for all cases but I do not see the proof for that.

src/optimizer/optimizer.py
```diff
@@ -613,13 +631,17 @@ def solve(self) -> Dict:
...
+            gap_abs = self.settings.gap_abs
+            solver = pulp.PULP_CBC_CMD(msg=0, threads=self.settings.num_threads,
+                                       timeLimit=self.settings.time_limit,
+                                       gapAbs=None if gap_abs is None else gap_abs * OBJECTIVE_SCALE)
```
ekkea: I guess this is the setting that actually has an effect on the performance due to the gapAbs setting.

ekkea (commented):
Here are my 2 cents w/o going into intensive testing: If this solves an urgent performance problem for us, I'm ok to apply the change, however, I think half of the changes here will have no predictable positive effect, the other half will have some performance effect but comes with the risk of not solving for proper strategy adherence.

andig:
Thanks, the objection about strategy loss is the load bearing one and it holds. I measured it rather than argue it, and it is worse than the PR body implies.

### The gap swallows the strategy terms on most cases

For every golden case, solved on this branch, the total weighted preference objective in currency units against `gap_abs = 0.01`:

| case | \|cost\| | \|3 · preference\| | gap / preference |
|---|---|---|---|
| 018-high-soc-initial | 162318 | 0.000067 | **149×** |
| 009-discharge-before-import | 3.48 | 0.000214 | **47×** |
| 010-infesible-charge-goal | 1397 | 0.000949 | **10×** |
| 023-c_min-limit-kept-with-p_demand-set | 0.74 | 0.001360 | **7×** |
| 011-infeasible-charge-demand | 2.52 | 0.001554 | **6×** |
| 012-early-charging-not-perfect | 5.17 | 0.002758 | **4×** |
| 015-low-soc-initial | 5253 | 0.003435 | **3×** |
| 019-unexpected-charge-spikes | 2.79 | 0.007683 | 1.3× |
| 017-battery-charge-priotization-2 | 11.30 | 0.016630 | 0.6× |
| 020-weird-charging-at-night | 0.35 | 0.025965 | 0.4× |
| 021-min-pv-use-case-with-weird-behavior | 8.55 | 0.038476 | 0.3× |
| 016-battery-charge-priotization-1 | 3.48 | 0.047610 | 0.2× |
| 013, 024-027 (peak and ramp) | — | 0.073 - 4.73 | ≪ 1 |

On eight of these the *entire* weighted preference objective is worth less than the absolute gap, several by one to two orders. The preference terms are one signed, so their total magnitude bounds their spread: on those cases every strategy variant sits inside the acceptance window and CBC returns whichever incumbent it happens to hold when the gap closes. The proof you asked for cannot be constructed there, it fails by arithmetic.

`OPTIMIZER_STRATEGY_WEIGHT` cannot close this. It buys a factor 3 against a shortfall of 3 to 150, and raising it runs into the cliff table in the description, 10× already gives up 0.4 percent of real money on 023. The weight and the gap work against each other by construction, not by tuning.

The peak and ramp cases survive because `prc_p_peak = penalty_base * 1e-3` is large enough to clear a cent. That is consistent with the weight being the only one of the two that helps peak shaping.

The second table in the description was already showing this and I read it too kindly: gap alone moves 15 to 28 MWh of charging away from the deliberate tie break on the flat cases. Weight 3 masks that to zero on five of six, but by branching order, not by construction.

### Related: the weight does nothing when prices are flat or negative

When `min_import_price` is at or below zero the tie breaker coefficients collapse toward zero, `3 × ~0` is still `~0`. Case 018 has `min_import_price = -9.9e-7` and its smallest scaled objective coefficients land at 1e-2 while the penalties sit at 1.4e6. That is exactly the cheap and flat tariff regime this PR targets.

### On the objective split

You are right that as it stands the split is routing for the weight and nothing more, and my own comment on it is stale, it still claims

```
# so solve() can optimize the two in sequence instead of in one near degenerate objective
```

which no longer happens anywhere. That is a leftover from the two stage version this branch is named after.

### Proposal

Restore the two stage solve, the split is already the infrastructure for it:

1. maximize `cost_objective` alone with `gapAbs` at one cent. This is where the speed is, the cost only model has none of the degenerate tie breaking that makes the plateau expensive.
2. add `cost_objective >= C* - budget`, where `budget` is the money we explicitly allow the preferences to spend.
3. maximize `preference_objective` alone, with a gap expressed in preference units.

Strategy adherence becomes something we solve for instead of something that emerges from coefficient ratios. No weight to tune, no cliff table, and the 0.1679 percent that `flat-0` gives up becomes a stated budget instead of a surprise. Cost is a second solve, which is heavily constrained by the cost bound, so it gets its own time limit and falls back to the stage one schedule on timeout. The degradation is then loss of preference, never loss of money.

If the second solve turns out too expensive, the cheaper single solve version is the one already floated in the description: normalize `preference_objective` so its total worth is a stated fraction of the cost objective, with that budget set *above* `gap_abs`. That is the minimum needed for the terms to survive the gap at all.

Either way this wants a test asserting `|weighted preference spread| > gap_abs` per case. Right now that invariant is violated on half the golden set and nothing notices.

### On `OBJECTIVE_SCALE`

Separate concern, it came in with #104, so better as its own issue. Measured coefficient ranges after scaling: realistic cases run 1e-2 to 1.4e6, dynamic range up to 1.4e8; the attenuate cases reach 3e8. A constant slides the window but cannot fix a range. The cheap principled version is to derive the factor per model so that `max|coef|` lands on a fixed target, same argmax invariance, no bet on price magnitude or units.

Happy to rework this along the two stage lines if that direction looks right to you.

andig:
@ekkea pushed the rework, the description is rewritten around it.

Short version of what your second comment turned into: you were right, and no weight could have fixed it. The two are now solved in sequence, cost first with the gap, preferences second under a constraint that keeps the money the first stage found. `OPTIMIZER_STRATEGY_WEIGHT` is gone with its cliff table, `OPTIMIZER_PREFERENCE_BUDGET` replaces it and defaults to zero.

Measured over 335 unique requests from the slow request dump, at the production limit of 10 s:

| variant | median | over 10 s | strategy better than main | strategy worse |
|---|---|---|---|---|
| main | 7.65 s | 149 | — | — |
| weight 3 + gap, what you reviewed | 0.72 s | 47 | 51 | **112** |
| two stage | 2.99 s | 52 | **120** | 79 |

The version you reviewed is behind main on strategy adherence on 112 of 335 production requests. That is your objection, on real traffic rather than on a constructed case. No golden case moves in either version, which is why neither of us saw it there.

On your first comment: the split is no longer only readability, it is what the second stage optimizes over. On the third, `OBJECTIVE_SCALE`, I left it out of this PR since it came in with #104 and stands on its own, and opened it separately.

Two things worth knowing beyond this PR, both found while benchmarking:

- PuLP's `warmStart=True` makes CBC return a strictly worse solution and report it as proven optimal, and declare a model infeasible over a bound the warm start itself satisfies. On request `4f28656db0ba`, preference -0.806 warm against -0.610 cold, the cold value matching a single joint solve exactly. Worth avoiding anywhere else in the codebase.
- Levelling a peak costs round trip efficiency, about 3.5e-5 currency on `024-attenuate-demand-peaks`, and the case refuses to level under a stricter cost bound. So the tier 3 peak and ramp terms are not the cost neutral tie breakers AGENTS.md describes. That is a modelling question rather than a solver one, happy to open it separately if you want it tracked.

andig:
The `OBJECTIVE_SCALE` comment is now #123, kept out of this PR since it belongs to #104.

Short version: there is no proof for the constant, and it does not place what it claims to. With `1e6` applied the largest coefficient in the model handed to the solver lands anywhere from 1.6e5 to 3e8 across the stored cases, six of nineteen above 1e6. #123 derives the factor from the model instead. On the slow request dump it measures as a wash, so it is a correctness and legibility change rather than a performance one, and closing it is a fine outcome if that is not worth the churn.

andig:
Current status brings p99 significantly down, but the second solve has negative impact on p50/p95:

Screenshot 2026-07-28 at 09 59 24

andig:
Pushed `4a15096`, which makes the split conditional. This is the answer to the cost the two stage solve was carrying on ordinary traffic, and it comes out of what production has been showing since the rollout at 07:04Z.

### The problem it fixes

Measured over 14 buckets on the live service, against the build before it:

| median across buckets | before | after the two stage solve |
|---|---|---|
| p50 | 0.22 s | 0.38 s |
| p90 | 0.80 s | 1.16 s |
| p95 | 1.56 s | 2.43 s |
| p99 | 10.19 s | **5.62 s** |

The tail did what it was meant to, and has not touched the limit in 14 buckets. The body got 70 percent slower, because the split spawns a second CBC on every request with a strategy configured, and the median request never needed it.

### What changed

Splitting only pays on a degenerate plateau. A request that proves the joint objective never had one: it decided its own tie, in one solve, with no gap and no preference budget to give anything away. So the solve now probes first, on a fifth of the time limit, and only a probe that runs out falls back to money-then-preferences.

On the stored cases, comparing `OPTIMIZER_PROBE_SECONDS=0` against the default in the same checkout, five runs each over two passes:

| | median case | total |
|---|---|---|
| main | 0.052 s | 2.19 s |
| this PR before the probe | 0.114 s | 3.55 s |
| **with the probe** | **0.053 s** | 1.76 s |

Back to main's median, and the total beats main because `020-weird-charging-at-night` gets routed *to* the split and finishes in 0.53 s against main's 0.93 s.

`OPTIMIZER_PROBE_SECONDS` tunes the budget, 0 restores the unconditional split.

### A pulp trap worth knowing about regardless of this PR

pulp reports `LpStatusOptimal` whenever CBC returns any feasible solution, including one it stopped on at the time limit. On a captured request a 2 s run and a 30 s run both said `Optimal`, with objectives of -682466848 and 59714881. The first version of the probe read `status` and so accepted every truncated search: it looked excellent, median 2.08 s with 331 of 335 requests apparently proving, because it was quietly cutting the whole population off at the probe budget. `sol_status == LpSolutionOptimal` is the only thing that means proven. Anything in this codebase that reads CBC status to mean "finished" is worth a second look.

### Known gap, not fixed here

One request of 335 captured slow ones comes back `Not Solved`: the probe finds no feasible solution at all and leaves the cost stage too little clock to find one either. A 1 s probe does not fix it. The probe's incumbent is kept whenever it has one, which covers every other case, but that request has nothing to keep. It needs a floor on the cost stage's budget before this is safe to rely on, and I would rather that were a separate change with its own measurement.

On the captured slow requests the probe is a net loss, which is what it should be on a population that is entirely hard: median 4.89 s against 3.03 s for the unconditional split, 214 of 335 falling through. Requests over the limit stay far below main, 51 against 149. The gain is on ordinary traffic, which that population by construction contains none of.

PR #130, opened 2026-08-01
Drop the step to step ramp from the leveling strategies

andig:
## What this is

`attenuate_*_peaks` priced two quantities: the horizon maximum, and the step to step ramp. This drops the ramp and leaves the peak alone.

The ramp is a sum over transitions, so every path that rises monotonically between the same two levels carries the same sum — it cannot tell a straight climb from a plateau. And it puts a second absolute value system on the same `p_grid` as the peak term, which is what costs the solve time: the extra rows are not the problem, the overlap is.

## Measurements

`uv run python tools/bench_peak_leveling.py [--rev main] [--strategy ...] [--sweep]` runs the stored cases, or the pinned peak sweep, through the working tree and a reference revision. It is added here so the numbers are reproducible.

All numbers: Apple M4, native CBC 2.10.13, best of 3 solves per case, against `main`.

### Solve time

| 20 stored cases | total |
|---|---|
| with the strategies the cases ship with | 2.069 s -> 2.069 s |
| forced `attenuate_grid_peaks` | 2.685 s -> **2.197 s (-18 %)** |

Only 5 of the 20 cases level anything with their own strategy, and the other 15 build a model the change does not touch — hence the flat first row.

### Model size

Per levelled side, `T - 1` variables and `2·(T - 1)` rows removed. 026 goes 117 -> 89 rows and 89 -> 75 columns.

### Peaks

Unchanged everywhere it matters:

- stored corpus, forced strategy: identical peak on 19 of 20 cases. The exception is 013, export 25123 -> 25818 W (+2.8 %), a real trade
- 486 pinned peak cases (`--sweep`, 3 x 6 x 3 x 3 x 3 = spike size x position x goal x charge power x base load): **peak raised in 0 of 486**
- largest cost delta over the sweep: 9e-8 currency

### What it gives up

The ramp did order the steps below a pinned maximum, and nothing replaces it. Stated plainly, because this is the actual price:

| 486 pinned peak cases, vs main | |
|---|---|
| profile changed | 432 of 486 (89 %) |
| deviation where changed | 1512 W -> 1810 W, mean **20 % higher** |
| cases where deviation got worse | 432 of 432 changed |
| peak raised | 0 of 486 |

One stored case moves with it: `027-attenuate-feedin-peaks-below-cap` had its export levelled to a flat 57.5 W below the 950 W cap by the ramp, and now takes 50 / 50 / 50 / 80 W. Same peak, same cost, less level — the stored schedule is updated to match rather than papered over.

Where the maximum is *not* pinned, the peak term alone still spreads charging: that is what `test_the_peak_term_spreads_the_charge_over_the_horizon` and cases 024 - 027 cover.

## Also here

- `tests/test_peak_leveling.py`: a guard against the ramp being reintroduced unnoticed, the peak variables existing only for the levelled sides, the spreading behaviour, and the blind spot above pinned as an explicit test so a future change to the term has to confront it
- `tools/bench_peak_leveling.py` with `--sweep`
- README and the objective term table updated for the removed term and the blind spot

## Note on what this PR was before

It originally added a third term — the distance of every step from a freely placed level — as the linear stand-in for a mean square deviation, to close exactly the blind spot above, and dropped the ramp as redundant next to it. That part is reverted.

The stored corpus did show the term working, but re-measuring against the production slow log (1184 requests using peak shaving strategies, 643 distinct) did not support the case:

- on `3d7f5648`, the request that looked like the clearest win, all three variants stop on the time limit with ~0.17 % gap still open. Its reported peak is whatever incumbent CBC held at cutoff, and it moves with the budget: at 30 s main 3497 W / with-deviation 1642 W / peak-only 1823 W, at 60 s 2080 / 2080 / **1823**
- on that request `penalty_base` is 3.866e-4, so the entire peak penalty for a 2080 W export peak is 8.0e-4 — against a remaining gap of 4.7e4. The levelling terms are ~7 orders of magnitude below anything the search reacts to, so they order nothing there either way

That does not disprove the deviation term, but it means the evidence offered for it was measuring solver noise, and the term is not worth carrying on that basis. The ramp removal is independent of all of it and is what remains.

---


ekkea:
I fully agree to removing the ramp - I do not see the benefits of the goal and it is making problems harder to solve for the optimizer.

Regarding the free-level approach, I see that performance is getting much worse for the "hard" cases. Few single tests indicate a factor for 3 and more. Also, we are adding quite a bit of complexity to the model.

Therefore, adding this needs to add value to the result of the optimization, i.e. there needs to be a somehow valuable goal behind that. I I'm struggling to see what that shall be - maybe I'm missing something here?

I'd suggest just to remove the ramp-goal and leave this new mechanism out unless there is really a value that it achives - which I do not see.

andig:
> Therefore, adding this needs to add value to the result of the optimization, i.e. there needs to be a somehow valuable goal behind that. I I'm struggling to see what that shall be - maybe I'm missing something here?

Not sure if I understand the question: The value is grid peak shaving and it's only active when the respective strategy is selected. Is this the question?

Issue #132, opened 2026-08-04
Solution to "reduce grid feedin peaks" not correct


ScumbagSteve:
I get this solution for "reduce grid feedin peaks" which seems implausible to me:

<img width="1349" height="684" alt="Image" src="https://github.com/user-attachments/assets/7ef375a0-e46f-4c85-bc53-cea1aa4ee258" />

The feedin peak it not avoided at all, it even seems it is doing the exact opposite (to maximize grid feedin peaks). 
The Tesla is charged by planner to 100% in the night, so ignore it not being considered for solar charging. This is plausible.

### Request:
```
{
  "batteries": [
    {
      "c_max": 10350,
      "c_min": 1380,
      "charge_from_grid": true,
      "d_max": 0,
      "p_a": 0.0001502226,
      "s_goal": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 54000, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0],
      "s_initial": 36180,
      "s_max": 54000,
      "s_min": 0
    },
    {
      "c_max": 2760,
      "c_min": 1380,
      "charge_from_grid": true,
      "d_max": 0,
      "p_a": 0.0001502226,
      "s_initial": 23750,
      "s_max": 25000,
      "s_min": 0
    },
    {
      "c_max": 4000,
      "c_min": 0,
      "charge_from_grid": true,
      "d_max": 4000,
      "discharge_to_grid": true,
      "p_a": 0.0001502226,
      "s_capacity": 7500,
      "s_initial": 735,
      "s_max": 6750,
      "s_min": 735
    }
  ],
  "eta_c": 0.9,
  "eta_d": 0.9,
  "grid": {},
  "strategy": {
    "charging_strategy": "attenuate_feedin_peaks",
[...]

PR #134, opened 2026-08-05
Reject an unknown strategy instead of silently dropping it

andig:
Found while analysing #132.

## What happens today

`Optimizer` matches the strategy names off the request against the branches that build the objective:

```python
self.peak_sides = PEAK_STRATEGY_SIDES.get(strategy.charging_strategy, ())
...
if self.strategy.charging_strategy == 'charge_before_export':
if self.strategy.discharging_strategy == 'discharge_before_import':
```

A name that matches none of them reaches no branch, so the request is solved without any strategy. Nothing in the response says so — the schedule is cost optimal for the request, `status` is `Optimal`, `objective_value` is right, and only the shape of the grid profile gives it away.

That is what a request answered like #132 looks like. Replaying that request on `main` with the strategy string intact levels grid export to a flat 658 W plateau; replaying it with the strategy misspelled reproduces the reported export peak of 1179.2 W exactly — the raw `max(ft - gt)` of the request, i.e. no levelling at all, at an identical bill.

`openapi.yaml` has carried the enum since the strategies were added:

```yaml
charging_strategy:
  enum: [none, charge_before_export, attenuate_demand_peaks, attenuate_feedin_peaks, attenuate_grid_peaks]
```

but the flask-restx input model declared a plain `fields.String`, so nothing validated it. The documented contract and the enforced one had drifted apart.

## The change

Declare the accepted names on the input model, listed in `optimizer.py` next to the code that reads them so the list and the branches cannot drift:

```python
CHARGING_STRATEGIES = ('none', 'charge_before_export', *PEAK_STRATEGY_SIDES)
DISCHARGING_STRATEGIES = ('none', 'discharge_before_import')
```

The existing `@api.expect(..., validate=True)` then does the rest, and names the offending value and the accepted set:

```json
{
  "message": "Input payload validation failed",
  "details": {
    "strategy.charging_strategy": "'attenuate_fedin_peaks' is not one of ['none', 'charge_before_export', 'attenuate_demand_peaks', 'attenuate_feedin_peaks', 'attenuate_grid_peaks']"
  }
}
```

An absent `strategy` object, an absent field, and `none` are all unchanged. A charging strategy passed in the discharging slot is now rejected too — the two sets are disjoint and were never interchangeable.

## Breaking change

A caller sending a name outside the documented enum gets a 400 where it used to get a schedule. That is the point: today it gets a schedule that quietly ignores what it asked for. Worth calling out for the release notes, since evcc is the caller.

## Verification

`uv run pytest` green, 98 tests. `uv run ruff check` clean. `tests/test_app.py::test_unknown_strategy_is_rejected` pins both halves: three rejected names, and every documented charging × discharging combination plus an absent strategy still solving.

## Not in scope

This does not change any schedule, and it is not a fix for the optimisation reported in #132 — that request levels correctly on current `main`. It removes the failure mode that makes a dropped strategy indistinguishable from a working one.

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -8,7 +8,6 @@ import pulp
 import pytest
 
 from optimizer.app import app, settings
-from optimizer.optimizer import CHARGING_STRATEGIES, DISCHARGING_STRATEGIES
 
 
 @pytest.mark.parametrize('test_case', pathlib.Path('test_cases').glob('*.json'))
@@ -155,4 +154,3 @@ def test_abort_returns_json_message():
     assert response.status_code == 400
     assert response.json is not None
     assert "message" in response.json
-
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -156,37 +156,3 @@ def test_abort_returns_json_message():
     assert response.json is not None
     assert "message" in response.json
 
-
-def test_unknown_strategy_is_rejected():
-    # a misspelled strategy reaches no branch in the optimizer, so the request used to be solved
-    # without any strategy and came back Optimal with a cost optimal schedule - nothing in the
-    # response says the strategy was dropped (#132)
-    client = app.test_client()
-    request = {
-        "batteries": [{
-            "s_min": 0, "s_max": 10000, "s_initial": 5000,
-            "c_min": 0, "c_max": 5000, "d_max": 5000, "p_a": 0.1,
-        }],
-        "time_series": {
-            "dt": [3600, 3600],
-            "gt": [1000, 1000],
-            "ft": [0, 0],
-            "p_N": [0.3, 0.3],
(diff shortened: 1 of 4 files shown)


andig:
@ekkea is this the correct approach for API validation?

tests/test_app.py
ekkea, line 1: The new test case should sit in a separate test file and also should test at least one negative test case.

ekkea (commented):
This is fine, just wondering what exactly is being improved. Invalid strategies were even rejected before as the openapi.yaml already had the enumerations defined.

tests/test_app.py
andig: The openapi spec was not enforced and a wrong strategy was silently ignored.

PR #135, opened 2026-08-07
Roll up release

andig:
Rollup branch off `main` combining:

- `perf/two-stage-solve` — two stage solve, cost objective first, preferences second
- #130 — level the grid profile against a free level instead of its peak and ramp

Merge conflicts resolved in two places, both from the same overlap: #130 rewrote the leveling terms that `perf/two-stage-solve` had moved into the preference objective.

- `src/optimizer/optimizer.py` — the peak and level deviation terms from #130 now add to `preference`, not `objective`, so they stay in the second stage.
- `docs/comparison_objective_terms.md` — kept both the tier 3 two stage paragraph and #130's ramp → level deviation wording.

Full test suite passes (124 passed).


andig:
All merged.

Issue #127, opened 2026-07-30
Minimum charge power not taken into account


CalvinSchwartz:
My wallbox/car (blue) has a minumum charge power of 4140W `"c_min": 4140`, however the optimizer recommends a charge power lower than that value.
Sometimes it works correctly and I get a response that takes the minimum charge into account, however on the next update it is not taken into account.

Request data is:
`{
  "batteries": [
    {
      "c_max": 11040,
      "c_min": 4140,
      "charge_from_grid": true,
      "d_max": 0,
      "p_a": 0.00030472202,
      "s_initial": 57488.586,
      "s_max": 62505,
      "s_min": 0
    },
    {
      "c_max": 5000,
      "c_min": 0,
      "d_max": 5000,
      "p_a": 0.00030472202,
      "s_capacity": 8000,
      "s_initial": 5840,
      "s_max": 7600,
      "s_min": 800
    }
  ],
  "eta_c": 0.9,
  "eta_d": 0.9,
  "grid": {},
  "strategy": {
    "charging_strategy": "charge_before_export",
    "discharging_strategy": "discharge_before_import"
  },
  "time_series": {REDACTED}
}`

<img width="821" height="428" alt="Image" src="https://github.com/user-attachments/assets/ec2c1935-8acd-481a-9078-34c396888357" />

andig:
Missing required request/response data

CalvinSchwartz:
Sorry for the late response, I had to wait until my car was empty enough. Now the same issue appears again, attached is the request and response data @andig 

[request.txt](https://github.com/user-attachments/files/30783955/request.txt)
[response.txt](https://github.com/user-attachments/files/30783954/response.txt)

andig:
@CalvinSchwartz are you using evcc optimizer or self-hosted?

andig:
Thanks for the data, that was enough to pin it down.

## Confirmed

Battery 0 (the EV, `c_min=4140`, `c_max=6900`, `d_max=0`) charges below its gate in 13 of the 192 slots of the attached response:

| slot | dt | charged (Wh) | gate = `c_min·dt/3600` |
|---|---|---|---|
| 0 | 385 s | 247.6 | 442.8 |
| 3 | 900 s | 615.5 | 1035.0 |
| 4–14 | 900 s | 382–604 | 1035.0 |

[...]

PR #136, opened 2026-08-07
Refuse a schedule whose binaries came back off the integers

andig:
Fixes the defect behind #127: an EV with `c_min=4140` charging at a fifth of that in 14 of 192 slots, `z_c` sitting between 0.12 and 0.36, reported as `Feasible`.

## What happens

`_solve_preferences` decides whether to keep its result by reading the variables back. A solve that runs out of clock before it finds an integer solution leaves the **relaxation** in them, and pulp reads that like any other result. It scores better on the preferences than any real schedule can — because it is one the model forbids — so `improved` is true, the first stage schedule is discarded, and `problem.status` is then forced to Optimal.

Reproduces on the split path at the default probe budget:

```
path=split  preference_stage=Not Solved  status=Feasible  5.70s
  slot   3: c=  615.46 gate= 1035.00  z_c=0.35678659
  ... 14 slots between 0.12 and 0.36
```

## What changes

**The acceptance test.** Integrality joins the two conditions already checked there, for the reason the comment beside them gives: a solver that reports the wrong status must not be able to spend money, and it must not be able to hand back a schedule the model does not allow either. The stage now ends as `Not Solved, kept the first stage` and the first stage schedule survives.

**A check before the result is built.** Standing in for every stage above deciding correctly. A schedule that breaks the model is worse than no schedule — it looks like an answer and the caller charges a battery by it — so a fractional point is reported as no schedule rather than returned. One pass over the integer variables against a solve measured in seconds. It logs when it fires; a silent fallback is how this stayed invisible.

Note for anyone reading the check: pulp stores a binary as an integer bounded to 0 and 1, so `LpBinary` never survives on a variable and `LpInteger` is what the filter has to match. Getting that wrong makes the guard pass vacuously.

## Tests

`tests/test_fractional_solution.py`, three cases, no captured request and no dependence on a timing race:

- a preference stage that returns without an integer solution is refused, first stage schedule intact
- a fractional point that reaches the result is reported as `Not Solved` with no schedule
- values inside CBC's own integer tolerance still pass, so the guard does not fire on rounding

Full suite passes (115).

## Base

Branched off `perf/two-stage-solve` rather than `main`, since that is where `_solve_preferences` lives. #135 picks it up on its next merge. Worth deploying independently of the perf work — production currently runs the rollup, where this is reachable.

Related: the probe change measured alongside this hides the bug by keeping most requests on the joint path. It does not remove it, and this fix stands on its own.

### DIFF
--- a/src/optimizer/optimizer.py
+++ b/src/optimizer/optimizer.py
@@ -80,6 +80,11 @@ COST_BOUND_SLACK = 1e-5
 # peak. A hundredth of a cent, three orders below the gap the cost stage may already leave.
 COST_BOUND_TOLERANCE = 1e-4
 
+# how far off 0 or 1 a binary may land and still count as integral. CBC's own integer tolerance
+# defaults to 1e-6 and it rounds within that before reporting, so anything above this came back
+# from a relaxation rather than from a rounding difference.
+INTEGRALITY_TOLERANCE = 1e-5
+
 # share of OPTIMIZER_TIME_LIMIT the probe gets before the solve falls back to the split. Kept well
 # under half: the probe is pure loss on a request that ends up splitting anyway, so it should be
 # long enough to catch the ordinary ones and no longer. Measured over the captured slow requests,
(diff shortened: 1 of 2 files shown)


andig:
@ekkea on top of the 2-stage solve. Merging since that hasn't hit master yet.

PR #138, opened 2026-08-08
feat: add a per step forced discharge demand

andig:
Follow-up to the optimizer TODO in evcc-io/evcc#31995 ("optimizer (MIP) integration to plan export windows from price spreads — deliberately out of scope; this PR is the mode + site-logic foundation"). That PR added the `api.BatteryDischarge` mode and `batteryGridDischargeLimit`; this one gives the optimizer a way to be told about it.

## API

New optional per-battery field `d_demand`, the mirror of the existing `p_demand`:

```jsonc
"d_demand": [0, 0, 2500, 2500, 0, 0]   // discharge energy per time step (Wh)
```

Added to the flask-restx model, the request parsing, the time-series length validation, `openapi.yaml`, and the generated Go client (`go generate ./...`).

## Model

- new penalty variable `d_demand_pen[i][t]`, only for batteries that carry a demand
- constraint `d[i][t] + d_demand_pen[i][t] >= min(d_max * dt/3600, d_demand[t])`
- penalty `-prc_p_goal_pen * d_demand_pen[i][t]` in the objective, no time weighting: a forced discharge names the steps it wants itself

Deliberately soft, like the charge demand it mirrors. A demand the battery cannot serve — already empty, `d_max` too small, the energy needed by the house — leaves an unserved remainder rather than turning the request infeasible. `prc_p_goal_pen` sits an order below `prc_soc_exc_pen`, so the forced discharge stops at the `s_min` reserve instead of draining through it, which is the same reserve guard `applyBatteryMode` enforces on the evcc side.

Reaching the grid still requires `discharge_to_grid`; the demand does not override it.

## Tests

`tests/test_force_discharge.py`: no demand leaves the battery alone, a demand is served against an unfavourable price spread, clipping to `d_max`, stopping at `s_min`, and staying solvable when there is no sink for the energy. Full suite green (103 passed), ruff clean.

### DIFF
--- a/client/client.gen.go
+++ b/client/client.gen.go
@@ -62,6 +62,13 @@ type BatteryConfig struct {
 	//   - False: (default) The battery cannot be charged while power is retrieved from grid
 	ChargeFromGrid bool `json:"charge_from_grid,omitempty"`
 
+	// DDemand Minimum discharge demand per time step (Wh). Forces the battery to discharge in the
+	// given steps, e.g. to sell into a high feed-in rate. The demand is clipped to d_max and
+	// is a soft goal: it is given up rather than draining the battery below s_min, and rather
+	// than making the request infeasible. Discharging into the grid additionally requires
+	// discharge_to_grid.
+	DDemand []float32 `json:"d_demand,omitempty"`
+
 	// DMax Maximum discharge power in W
 	DMax float32 `json:"d_max"`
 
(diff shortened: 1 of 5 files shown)


ekkea (changes requested):
two things missing:

- missing check: p_demand and d_demand being >0 at the same time needs to be closed out. What shall happen? Reject and send 500?
-  there is no precaution that d_demand may run into SOC becoming 0. Keeping up the penalty once this happens will lead to weird behavior as we have seen this with p_demand previously.
- i strongly recommend to add standard test cases (in form of requests that use this feature, added to the test_cases folder). They are much easier to review than specific test scripts.

